PR #2, opened 2016-03-21
First Part of Selenium Driver for NDT HTML5 Client

rowanv:
This PR includes the following: 
1. Initial implementation of the test sequence with the Selenium driver
2. Error handling for test timeout, with corresponding unit tests
3. Error handling for failure to load the landing page, with corresponding UTs,
4. Error handling for missing numeric values, with corresponding UTs.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/2)


mtlynch:
Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[.travis.yml, line 10 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOuhzFFKHkeirSR5I9:-KDOuhzFFKHkeirSR5IA:1749568530) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/.travis.yml#L10)):*
I'd like to eliminate any dependency on firefox/selenium for unit tests. Can we mock out the calls to selenium in the unit tests that require travis to have an x server?

---

*[requirements.txt, line 11 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOu6v4PcKRQLluhbW4:-KDOu6v5hadpMcOdHs_W:180378181) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/requirements.txt#L11)):*
Can we add the libs as we need them. It doesn't seem like we're actually using several of these:

* coverage
* docopt
* funcsigs
* pbr
* requests
* six

---

*[requirements.txt, line 12 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOtO1iPOOlv82YYeQI:-KDOtO1iPOOlv82YYeQJ:-984304606) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/requirements.txt#L12)):*
Can we split these into:

1. What an end-user needs to run the application (without running any tests)
1. What additional libs beyond (1) a developer needs to run tests
1. What additional libs beyond (1) and (2) that Travis needs to do its tests

This is what I have in mind:

https://github.com/m-lab/bigsanity/blob/master/.travis.yml

---

*[client_wrapper/client_wrapper.py, line 24 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOmvk-soabKJv1AcBW:-KDOmvk-soabKJv1AcBX:1138512644) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L24)):*
Can we always use double quotes for triple-quoted strings?

---

*[client_wrapper/client_wrapper.py, line 41 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOnSGgt_Uwcvv2TD8T:-KDOnSGgt_Uwcvv2TD8U:1173498514) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L41)):*
Can we reflow comments so that they break at 80 characters (max line length = 80)? I think this is my fault because I wrote the spec in Google Docs and was just eyeballing line breaks.

---

*[client_wrapper/client_wrapper.py, line 45 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOogCj-BP6UHGl8-8C:-KDOogCj-BP6UHGl8-8D:-1261567663) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L45)):*
Sorry, I realize this is a spec bug. `NdtResult` should actually contain `NdtSingleResult` objects instead of replicating their fields. I've updated the spec:
https://docs.google.com/document/d/17NPKEexO3FpP9K4mNfxbrGAiOiGznuk_7NPJzA5pN6I/edit#

Can we update the code to match? Sorry for the error.

---

*[client_wrapper/client_wrapper.py, line 50 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOo4aOfc88eqS3139-:-KDOo4aOfc88eqS31390:372816741) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L50)):*
Can we keep speed always in units of Mbps? Also, I'd like to stick to "c2s"/"s2c" in code instead of "upload"/"download" (though prefer "upload (c2s)" in comments/documentation) and "throughput" instead of "speed"

---

*[client_wrapper/client_wrapper.py, line 72 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOwl8vDnnpgcBjtoXK:-KDOwl8vDnnpgcBjtoXL:1171351740) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L72)):*
These should be assigned unconditionally. If `latency` is `None`, `NdtResult` should have a `latency` attribute with the value `None` rather than no `latency` attribute.

---

*[client_wrapper/client_wrapper.py, line 93 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOqmrmP4YJC5xXkeQ5:-KDOqmrnlrre-XN2Y-iu:-16190930) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L93)):*
Be mindful of functions and attributes that shouldn't be part of the class's public API. Prefix them with an underscore to mark them as protected (e.g. `self._result_values`)
https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming

---

*[client_wrapper/client_wrapper.py, line 100 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOnABvtSoIrS2k9zsN:-KDOnABvtSoIrS2k9zsO:-1473510752) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L100)):*
One line summary should be on same line as quote characters
https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

---

*[client_wrapper/client_wrapper.py, line 106 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOnGX0ibxUy2QhMsDz:-KDOnGX1tCOjGIMQRTqw:-948041256) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L106)):*
Need an extra line break before and after "Returns:"
https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

---

*[client_wrapper/client_wrapper.py, line 150 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOtBLzqsJEn8ovxZs_:-KDOtBLzqsJEn8ovxZsa:-1818594860) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L150)):*
`timeout_time` is a good idea. Can we document it?

---

*[client_wrapper/client_wrapper.py, line 155 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOrRuAt1n0eKq2aalF:-KDOrRuAt1n0eKq2aalG:2091434868) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L155)):*
If we have instance variables, we should document them in the class docstring and set them to `None` in `__init__`

That said, I'd like us to avoid instance variables for `NdtHtml5SeleniumDriver` if possible. It doesn't seem logical that it would have internal state, since if a caller does something like this:

```
driver = NdtHtml5SeleniumDriver()
a = driver.perform_tests('ndt.foo.com', 'firefox')
b = driver.perform_tests('ndt.bar.com', 'firefox')
```

The driver shouldn't be carrying state over from the first call to the second.

I think we can make the code more readable if, instead of passing around data implicitly through instance variables, we pass data around explicitly through function parameters.

Also look for opportunities to convert class functions into module-level functions  if they don't depend on any of the class's internal state. For example, `set_test_browser` can be a module-level function outside of the class.

---

*[client_wrapper/client_wrapper.py, line 216 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOuashvLksuBUZxc9L:-KDOuashvLksuBUZxc9M:-1991908940) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L216)):*
This is a fairly long function. Can we break it up into smaller functions?

---

*[tests/test_client_wrapper.py, line 30 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOv7UCYxZRp2eAS0_b:-KDOv7UDBjY69aGIl9tL:1129961725) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/tests/test_client_wrapper.py#L30)):*
I'd like to mock out selenium rather than functions in client_wrapper.

---

*[tests/test_client_wrapper.py, line 47 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOvgiDtM6ClqQSV-zX:-KDOvgiDtM6ClqQSV-zY:-274538626) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/tests/test_client_wrapper.py#L47)):*
This indenting due to the `with`s is a bit unfortunate. Can a pattern like this work here:
https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/tests/test_update.py

Alternatively consider doing the `patch` as an `@` decorator on the function

---

*[tests/test_client_wrapper.py, line 75 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOvuKzYhtWb6UY6GyL:-KDOvuKzYhtWb6UY6GyM:954269869) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/tests/test_client_wrapper.py#L75)):*
In general, can we use `mock.patch.object` so we get better static checking?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

rowanv:
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[.travis.yml, line 10 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOuhzFFKHkeirSR5I9:-KDYOUpoJlgkiStoSZdy:1353626339) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/.travis.yml#L10)):*
Sounds good, changed the tests so we now mock out all calls to selenium.

---

*[requirements.txt, line 11 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOu6v4PcKRQLluhbW4:-KDSeeeAMcf2LRzdISPi:-1758465522) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/requirements.txt#L11)):*
Hi Michael, 

six, pbr, and funcsigs are prerequisites for mock. requests, docopt, and coverage are prerequisites for coveralls, which is used by Travis CI, but am removing the coveralls dependencies.

---

*[requirements.txt, line 12 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOtO1iPOOlv82YYeQI:-KDShncjKmWeu017uYmD:1406448132) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/requirements.txt#L12)):*
Sounds great, fixed.

---

*[client_wrapper/client_wrapper.py, line 24 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOmvk-soabKJv1AcBW:-KDSiPQ2dc7K0zYOijyL:-1078048611) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L24)):*
Thanks, fixed all instances.

---

*[client_wrapper/client_wrapper.py, line 41 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOnSGgt_Uwcvv2TD8T:-KDSk3gSVONfMD_ZocEh:391433769) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L41)):*
Sounds good, checked and fixed all of the comments. 

---

*[client_wrapper/client_wrapper.py, line 45 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOogCj-BP6UHGl8-8C:-KDSnW0oesHX3rs0eP2I:611365756) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L45)):*
Not a problem, fixed.

---

*[client_wrapper/client_wrapper.py, line 50 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOo4aOfc88eqS3139-:-KDSqAD4w8FPMS1Rw6fT:-1396008396) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L50)):*
Yup! Went through and changed upload/download/speed in the tests and the original `client_wrapper.py` file. 

---

*[client_wrapper/client_wrapper.py, line 72 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOwl8vDnnpgcBjtoXK:-KDSk_J_sgKiLj-jbCua:951336568) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L72)):*
Sounds good. Fixed for latency, c2s and s2c.

---

*[client_wrapper/client_wrapper.py, line 93 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOqmrmP4YJC5xXkeQ5:-KDSr795qA7ACqmkQvyZ:-2003896356) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L93)):*
Thanks, fixed.

---

*[client_wrapper/client_wrapper.py, line 100 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOnABvtSoIrS2k9zsN:-KDSlMo5ZxKwWpbKSgJG:-446521175) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L100)):*
Thanks. Checked all comments and fixed. 

---

*[client_wrapper/client_wrapper.py, line 106 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOnGX0ibxUy2QhMsDz:-KDSla_fmYp3q-AP_sYn:-2003896356) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L106)):*
Thanks, fixed.

---

*[client_wrapper/client_wrapper.py, line 150 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOtBLzqsJEn8ovxZs_:-KDSm8vaJmcvw9NNNSI6:-1546924208) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L150)):*
Yes, fixed.

---

*[client_wrapper/client_wrapper.py, line 155 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOrRuAt1n0eKq2aalF:-KDXYwaldy-HccR2DOBK:1446486109) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L155)):*
That makes sense. Changed it so `driver` is no longer an instance variable and is passed around explicitly,  `set_test_browser` is now a  module-level function.

---

*[client_wrapper/client_wrapper.py, line 216 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOuashvLksuBUZxc9L:-KDXaYMYXI12F5TqOGTv:1287414153) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L216)):*
Yes, definitely. Split it up so it now calls `record_test_in_progress_values` and `populate_metric_values`.

---

*[tests/test_client_wrapper.py, line 30 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOv7UCYxZRp2eAS0_b:-KDYOPnPOXJ62hrQUY7O:-1152593692) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/tests/test_client_wrapper.py#L30)):*
Sounds good, changed to mock out selenium calls.

---

*[tests/test_client_wrapper.py, line 47 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOvgiDtM6ClqQSV-zX:-KDYNwNEtBGRqV_hhK3T:1305497434) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/tests/test_client_wrapper.py#L47)):*
Yes, that definitely looks better. Went ahead and changed all instances to follow that pattern.

---

*[tests/test_client_wrapper.py, line 75 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOvuKzYhtWb6UY6GyL:-KDYNDdoaWQl3-Jx_O2L:1958583599) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/tests/test_client_wrapper.py#L75)):*
Yes, fixed all instances. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

mtlynch:
Review status: 2 of 7 files reviewed at latest revision, 21 unresolved discussions.

---

*[test_requirements.txt, line 4 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYuRdowqI93kMsHaWb:-KDYuRdowqI93kMsHaWc:1125657840) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/test_requirements.txt#L4)):*
Won't pip install all of mock's pre-requisites automatically? If so, I'd rather not specify them explicitly.

---

*[client_wrapper/client_wrapper.py, line 45 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOogCj-BP6UHGl8-8C:-KDYxbAjM6cQD5_51tic:1501749548) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L45)):*
Can you review the design doc again? It doesn't look like `NdtSingleResult` is implemented in this PR and the documentation calls `c2s_start_time` an `NdtSingleResult` though the implementation seems to treat it as a `datetime`. 

Let me know if I can clarify anything.

---

*[client_wrapper/client_wrapper.py, line 20 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYutin1PNWDS39X9TB:-KDYutioLLWNUqzj9kTF:1424056143) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L20)):*
Can we change this so that we're only importing packages/modules?

>Use imports for packages and modules only.

https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports

This applies to `datetime`, `WebDriverException`, and `TimeoutException`. Is `expected_conditions` a module or a class?

---

*[client_wrapper/client_wrapper.py, line 85 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYwBbfjl-K64oB5pxH:-KDYwBbfjl-K64oB5pxI:-881614363) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L85)):*
Similar to `self.driver`, can we get rid of this as an instance variable and pass these variables around explicitly with function parameters? I'd like `NdtHtml5SeleniumDriver` to be stateless if possible.

---

*[client_wrapper/client_wrapper.py, line 91 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYz1Dyn7CDA7iFYO0g:-KDYz1Dyn7CDA7iFYO0h:-7252787) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L91)):*
`perform_test` should be `NdtHtml5SeleniumDriver`'s only public method, so the others should be prefixed with an underscore to indicate that they're protected. Also, if we get rid of `self._result_values` these functions can all become module-level functions outside the class.

---

*[client_wrapper/client_wrapper.py, line 101 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ0qlYmDPFlTGLBG3U:-KDZ0qlYmDPFlTGLBG3V:-2010657158) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L101)):*
I'd like to restructure the logic in this class a bit.

I think we should be validating data as close to possible as when we collect it. The current implementation is structured to collect data first, then defer checking until later. I'd prefer if it was more like:

```
_get_c2s_throughput(driver):
  upload_speed = driver.find_element_by_id('upload-speed').text
  if not upload_speed:
    raise ElementNotFoundError('expected page element was not found: upload-speed')
  try:
    return float(upload_speed)
  except ValueError:
    raise IllegalValueError('illegal value shown for c2s throughput: %s' % upload_speed)
```

The code will repeat a lot for latency, s2c, c2s, so this would probably be refactored into something more generic, but I hope the idea is clear.

Does that make sense?

---

*[client_wrapper/client_wrapper.py, line 165 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ2lZzPxI6Im5QUsPL:-KDZ2lZzPxI6Im5QUsPM:38483703) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L165)):*
These three chunks are very similar. Can we refactor to a more generic function to reduce code repetition?

---

*[client_wrapper/client_wrapper.py, line 167 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ2wAEXixny-FBqlua:-KDZ2wAEXixny-FBqlub:1231000068) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L167)):*
I'd prefer not to offer a default value for `timeout_time`

---

*[client_wrapper/client_wrapper.py, line 176 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYw0dWf1Q_eg9VLPhM:-KDYw0dWf1Q_eg9VLPhN:1646993351) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L176)):*
Suggest simplifying name to `timeout`

---

*[client_wrapper/client_wrapper.py, line 186 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ3EAalMu6NUI5VdDX:-KDZ35RhUTdSu89Sz8eI:-168543426) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L186)):*
Why are we checking for this error in particular? Can we just handle any error at this stage in the same way?

---

*[client_wrapper/client_wrapper.py, line 189 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ3PcmXBJ28rrho35W:-KDZ3PcmXBJ28rrho35X:785642305) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L189)):*
Suggest using `contextlib.closing` so we don't have to keep remembering to close `driver` before exiting scope.

---

*[client_wrapper/client_wrapper.py, line 216 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYydz6O1j6m4QCC7bX:-KDYydz7jwl_ODWZPpDg:-311536794) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L216)):*
Suggest renaming to `_create_browser` since it doesn't really set any state now.

---

*[client_wrapper/client_wrapper.py, line 226 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYyO2y9Fx7eFmKAVoO:-KDYyO2y9Fx7eFmKAVoP:1630679264) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L226)):*
Can we raise a `ValueError` if the caller specifies an unrecognized browser and `NotImplementedError` if they specify one that isn't yet supported?

---

*[tests/test_client_wrapper.py, line 17 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ45N4x2KfZG0M8LSx:-KDZ45N4x2KfZG0M8LSy:-422038571) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L17)):*
`client_wrapper` should be last

>Imports should be grouped with the order being most generic to least generic:
* standard library imports
* third-party imports
* application-specific imports

https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#Imports_formatting

Also, same notes about `import`s from client_wrapper.py apply here.

---

*[tests/test_client_wrapper.py, line 22 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6l0lu58WcmxFFgeY:-KDZ6l0lu58WcmxFFgeZ:1357711466) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L22)):*
Can we rename these to `NdtHtml5SeleniumDriver*Test`

Also, adding additional `TestCase` objects is okay, but not ideal. If multiple tests can use the same set of mocks (the instance members), I'd prefer to combine into a single `TestCase` class where possible.

---

*[tests/test_client_wrapper.py, line 71 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ5gIV5tBA5NkZfaVt:-KDZ5gIV5tBA5NkZfaVu:-762093684) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L71)):*
Test functions follow same conventions as regular functions, so we need to start with a one-line docstring.

---

*[tests/test_client_wrapper.py, line 79 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ58r9oLAWBJ0Hrs8f:-KDZ58r9oLAWBJ0Hrs8g:-971255816) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L79)):*
A lot of these mock classes seem like they could be simplified by using `mock.Mock`

For example, I think we could replace this class with:
`mock_web_element = mock.Mock(text='Non-numeric value')`

---

*[tests/test_client_wrapper.py, line 124 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ8gSecL77cteeXQya:-KDZ8gSf1rI97dqPULRG:-1969846996) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L124)):*
I'd like to avoid requiring test code to have to remember to call `stop()`. To clarify, I'm not against `with` context handlers, but I'd like to explore alternatives so we don't end up indenting like crazy. A single `with` is fine, so that's probably appropriate here.

---

*[tests/test_client_wrapper.py, line 164 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ61hIM0oRL2gTRVR_:-KDZ61hIM0oRL2gTRVRa:902389315) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L164)):*
Can we use obviously dummy URLs to make it obvious it's not hitting a real server? e.g. `http://ndt.mock-server.com:7123`

---

*[tests/test_client_wrapper.py, line 204 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6VAG5dTTaM7eEfLw:-KDZ6VAH8Jth-h_LIYRR:1213788308) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L204)):*
Can we include a test that checks that the timestamp is correct? (mocking out calls to the system clock)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

mtlynch:
Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

rowanv:
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[test_requirements.txt, line 4 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYuRdowqI93kMsHaWb:-KDbaTH1oYy-uStK6JmC:-1574287476) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/test_requirements.txt#L4)):*
Ah yes, sure. Went ahead and removed them.

---

*[client_wrapper/client_wrapper.py, line 45 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOogCj-BP6UHGl8-8C:-KDcK1W1VtP297jqW7Yi:-745462452) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L45)):*
Ah, yes, I hadn't implemented it yet since it looked like I was going to be exceeding 300 lines by quite a bit if I did, so I thought I'd go ahead and get feedback on work so far. Let me know if I should go ahead and include it in this wave of changes, thanks.

---

*[client_wrapper/client_wrapper.py, line 20 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYutin1PNWDS39X9TB:-KDc2p6D7mVdzZHPcIud:333499132) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L20)):*
Thanks, fixed. `expected_conditions` is a module, so I left it in. 

---

*[client_wrapper/client_wrapper.py, line 85 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYwBbfjl-K64oB5pxH:-KDbgyzwURymhUrJfXHU:1718834792) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L85)):*
Sure, fixed.

---

*[client_wrapper/client_wrapper.py, line 91 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYz1Dyn7CDA7iFYO0g:-KDbh2CG0R_xx7Z2P2lK:-1962606206) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L91)):*
Sounds good, fixed.

---

*[client_wrapper/client_wrapper.py, line 101 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ0qlYmDPFlTGLBG3U:-KDcIqplntwhdft1ytVX:1351848230) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L101)):*
Sure, that sounds good. I did not end up raising an error when an element is not found, since Selenium does that automatically. However, I refactored the class so that the check for metric validity happens right after Selenium reads in the element value. Let me know if I should change it further, thanks,

---

*[client_wrapper/client_wrapper.py, line 165 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ2lZzPxI6Im5QUsPL:-KDbxXntKLGC9hj3-GN3:-579954794) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L165)):*
Sounds good, I refactored it. Looks more readable now as well. 

---

*[client_wrapper/client_wrapper.py, line 167 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ2wAEXixny-FBqlua:-KDbhNcOzQEyMjaddVfA:-1962606206) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L167)):*
Sounds good, fixed.

---

*[client_wrapper/client_wrapper.py, line 176 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYw0dWf1Q_eg9VLPhM:-KDbhS4-1ttC3YsYbpoT:-1962606206) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L176)):*
Sounds good, fixed.

---

*[client_wrapper/client_wrapper.py, line 186 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ3EAalMu6NUI5VdDX:-KDcGIL-rtRpTzY19Gae:-1353835107) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L186)):*
Sounds good, changed it so we handle any error at this stage in the same way, updated unit tests.

---

*[client_wrapper/client_wrapper.py, line 189 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ3PcmXBJ28rrho35W:-KDby_9Yi7JEosVnyq2G:315396943) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L189)):*
Good suggestion, thanks.

---

*[client_wrapper/client_wrapper.py, line 216 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYydz6O1j6m4QCC7bX:-KDbl4W0fdtHIK8pKOam:-1962606206) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L216)):*
Sounds good, fixed.

---

*[client_wrapper/client_wrapper.py, line 226 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDYyO2y9Fx7eFmKAVoO:-KDbmr-RZNc2gWtQLT9L:1476777303) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/client_wrapper/client_wrapper.py#L226)):*
Sounds good, added with unit tests.

---

*[tests/test_client_wrapper.py, line 17 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ45N4x2KfZG0M8LSx:-KDc91KRV8Yf5bolDemp:-1991244860) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L17)):*
Thanks, fixed. 

---

*[tests/test_client_wrapper.py, line 22 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6l0lu58WcmxFFgeY:-KDcF5OmMH4RrvrFipnV:925850170) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L22)):*
Sounds good, fixed the naming. I went back through and tried to consolidate further, but given that all 3 classes have slightly different behaviour for the mock `driver`, I don't think it's possible to consolidate them further, unless we want to repeat the mock `driver` creation in each unit test. Let me know what you think would make the most sense, thanks,

---

*[tests/test_client_wrapper.py, line 71 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ5gIV5tBA5NkZfaVt:-KDboGFvn1MuFvwpRGZR:-1962606206) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L71)):*
Sounds good, fixed.

---

*[tests/test_client_wrapper.py, line 79 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ58r9oLAWBJ0Hrs8f:-KDcBLz0BL-gfsEkmewd:-87817201) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L79)):*
Right, so there were a couple of reasons I ended up explicitly creating the mock classes. With `NewWebElement`, I found that when using `mock.Mock`, `NewDriver`'s child mocks would override my desired behaviour for `WebElement`. The other reason was that with `NewDriver`, I'd like to eventually be able to test non-numeric metrics for `latency`, `c2s`, and `s2c` separately, which seemed easier to handle via an explicit `NewDriver` `find_element_by_id` method than by using what the `mock` documentation suggests, since that involves passing a list of all of each of the different values that the attribute (`NewWebElement.text`) should take throughout the life of the test. Since WebElement instances are created a bunch of times (and several of the instances have nothing to do with this test), this seemed clearer than the alternative, and all tests wouldn't then have to be changed if an extra `WebElement` instance ended up occurring at a future date in the original `client_wrapper.py`. Let me know if there's a solution that you think would be clearer, though. 

---

*[tests/test_client_wrapper.py, line 124 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ8gSecL77cteeXQya:-KDcAIkcb5BeONA_9CMu:1180323539) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L124)):*
Sounds good. Changed all mocking instances where necessary so it's no longer necessary to remember to call `stop()`.

---

*[tests/test_client_wrapper.py, line 164 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ61hIM0oRL2gTRVR_:-KDboVm_u1YwKhUG3byW:-1546924208) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L164)):*
Yes, fixed.

---

*[tests/test_client_wrapper.py, line 204 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6VAG5dTTaM7eEfLw:-KDc0QhNhhlNWfsO89Jb:-2110965386) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L204)):*
Sounds good, added.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

mtlynch:
Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[client_wrapper/client_wrapper.py, line 45 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDOogCj-BP6UHGl8-8C:-KDcYBTs6CIGEX2XDwjo:2052356559) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/9ffb60e93487267715c268b82eca059d44032ed3/client_wrapper/client_wrapper.py#L45)):*
That makes sense.

---

*[client_wrapper/client_wrapper.py, line 93 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcPiLiWxvNvoTghIbE:-KDcPiLjzFbQjFQLHiZD:670976017) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L93)):*
(nitpick) Missing trailing period.

---

*[client_wrapper/client_wrapper.py, line 95 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcPod65UAjF7eXblcE:-KDcPod7QSy6tjLXwB4I:1853964576) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L95)):*
Would prefer to use the `NdtResult` object from the start rather than converting it from a dictionary just before returning. Also, local variables / parameters should never have a `_` prefix.

---

*[client_wrapper/client_wrapper.py, line 111 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcTYPFXFB7u-YmoPtf:-KDcTYPGAPCwxLFlSbju:-1327028516) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L111)):*
It feels like we're mixing layers of abstraction. I'd prefer to keep`perform_test` very high level and delegate low-level tasks (like finding page elements by xpath) to sub-functions.

Can we rewrite the logic so that it looks kind of like this:

```
result = NdtResult()

if not _load_url(driver, url, result.errors):
  return result

if not _click_start_test_button(driver, result.errors):
  return result
result.start_time = _utcnow()

if not _wait_for_upload_test_start(driver, result.errors):
  return result
result.s2c_start_time = _utcnow()

if not _wait_for_c2s_test_start(driver, result.errors):
  return result
result.c2s_start_time = _utcnow()

if not _wait_for_results_page(driver, result.errors):
  return result
result.end_time = _utcnow()

parsed_results_page = _parse_results_page(driver)
result.errors.extend(parsed_results_page['errors'])
result.c2s_throughput = parsed_results_page['c2s_throughput']
result.s2c_throughput = parsed_results_page['s2c_throughput']
result.latency = parsed_results_page['latency']
return result
```

---

*[client_wrapper/client_wrapper.py, line 143 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcQGctrUmNwDpbIowk:-KDcQGctrUmNwDpbIowl:1618655266) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L143)):*
Can we add a message to the `ValueError` to help with diagnosis in case this ever bubbles up to the user, e.g. `ValueError('invalid browser specified: %s' % browser)`

---

*[client_wrapper/client_wrapper.py, line 199 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcZPEMMIPggmIJyzgS:-KDcZPEMMIPggmIJyzgT:241586428) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L199)):*
We should omit backticks for variable names in comments.

---

*[client_wrapper/client_wrapper.py, line 202 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcZJn5990lYrAPH4kh:-KDcZJn5990lYrAPH4ki:974328654) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L202)):*
>80 columns

---

*[client_wrapper/client_wrapper.py, line 221 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDccPh_I78Y5QbwWAlE:-KDccPh_I78Y5QbwWAlF:-1663956016) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L221)):*
For return value, we should describe in prose rather than the variable name.

---

*[client_wrapper/client_wrapper.py, line 236 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcZtBV4Yn8cmy2LoPZ:-KDcZtBV4Yn8cmy2LoP_:922366709) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L236)):*
I prefer abbreviations when we can avoid them. `_validate_metric`?

---

*[client_wrapper/client_wrapper.py, line 245 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDc_-O3gOu0iOYgtY1M:-KDc_-O3gOu0iOYgtY1N:-895209206) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L245)):*
Suggest just `the name of the metric to validate`. The function doesn't really care what the user puts in here, so we're better off not being overly prescriptive because then the docstring goes out of date if we want to support other metrics.

---

*[tests/test_client_wrapper.py, line 22 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6l0lu58WcmxFFgeY:-KDcc0_MYp2cJLG6ssiD:-1197841198) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L22)):*
>all 3 classes have slightly different behaviour for the mock driver

Can you elaborate a bit on this? I'm wondering if we can solve this by creating all the mocks in the `setUp` method but defining the test-specific logic for the mocks in the test methods themselves.

---

*[tests/test_client_wrapper.py, line 79 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ58r9oLAWBJ0Hrs8f:-KDcaUh6fMio9sdVpyV1:-555761673) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L79)):*
That sounds reasonable. Your explanation makes sense.

---

*[tests/test_client_wrapper.py, line 82 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDc_UN1GljFuoteTG6v:-KDc_UN28GmeNJCoHv6Q:1526587491) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/tests/test_client_wrapper.py#L82)):*
Neat, I've never seen this library before. Cool name.

---

*[tests/test_client_wrapper.py, line 107 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDc_skw4PfVJKccz2YY:-KDc_skw4PfVJKccz2YZ:-877286819) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/tests/test_client_wrapper.py#L107)):*
Can we test this a bit more rigorously? I'd like to know that all the recorded times are correct when time is advancing.

One thought is to patch the `now` function so that we define its sequence of return values.

---

*[tests/test_client_wrapper.py, line 175 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcajFRgUhpNES5Rxq9:-KDcajFRgUhpNES5RxqA:-1771102985) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/tests/test_client_wrapper.py#L175)):*
For single-line docstrings, closing triple quote should be on same line.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

rowanv:
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[client_wrapper/client_wrapper.py, line 93 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcPiLiWxvNvoTghIbE:-KDcqingpbdF8STjMIgh:-2003896356) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L93)):*
Thanks, fixed.

---

*[client_wrapper/client_wrapper.py, line 95 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcPod65UAjF7eXblcE:-KDdEvDgV2i3BkgzYSLj:1874996301) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L95)):*
Sounds good. Fixed so uses `NdtResult` object throughout, removed the `_` prefix.

---

*[client_wrapper/client_wrapper.py, line 111 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcTYPFXFB7u-YmoPtf:-KDdGUsxNwbC6Ga8GBhy:-1412990965) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L111)):*
Ah, ok. Sure, I went ahead and moved the click start test button logic into its own function. With regards to the other steps for the test, though, I wanted to make sure I'm understanding the general idea --  at each step we are catching errors and adding them to `result.errors`, but their presence triggers a return of `perform_test` in these remaining (the `WebDriverException` and the `TimeoutException`). If we move the try/except logic into a sub-function that undertakes part of the testing process, then each of these sub-functions would also need to return a variable that would indicate whether or not that sub-function had been successful/the larger `perform_test` should return its results? Or am I misunderstanding how that logic would work? Thanks.

---

*[client_wrapper/client_wrapper.py, line 143 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcQGctrUmNwDpbIowk:-KDcrT-y_15itpE88NeU:1718834792) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L143)):*
Sure, fixed.

---

*[client_wrapper/client_wrapper.py, line 199 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcZPEMMIPggmIJyzgS:-KDcrqmLX6_KKjZ9suBo:-374393068) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L199)):*
Sure, removed backticks here and in other places where they surrounded variable names in comments.

---

*[client_wrapper/client_wrapper.py, line 202 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcZJn5990lYrAPH4kh:-KDcs6wxpi2x0ECLlseH:-2003896356) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L202)):*
Thanks, fixed.

---

*[client_wrapper/client_wrapper.py, line 221 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDccPh_I78Y5QbwWAlE:-KDcsGS3ThxCG73KDIlT:-1962606206) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L221)):*
Sounds good, fixed.

---

*[client_wrapper/client_wrapper.py, line 236 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcZtBV4Yn8cmy2LoPZ:-KDcsccwT53vxoAm38wH:1250065474) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L236)):*
Sounds good, changed.

---

*[client_wrapper/client_wrapper.py, line 245 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDc_-O3gOu0iOYgtY1M:-KDcsoQ_Z31Uro5bQseZ:-1536417447) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L245)):*
sounds good, changed docstring.

---

*[tests/test_client_wrapper.py, line 22 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6l0lu58WcmxFFgeY:-KDdJPyBKSStfrmKWGCI:-1454966916) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L22)):*
Sure, so the most clearcut example is the difference between `NdtHtml5SeleniumDriverInvalidURLTest` and `NdtHtml5SeleniumDriverGeneralTest`. The former has our mock driver act like a normal driver throughout our application, then the second just has our mock driver is throw a `WebDriverException` as a side effect when it's first instantiated. So if we just patch everything by patching `client_wrapper.webdriver.Firefox`, then maybe we could just change `self.mock_driver.return_value = NewDriver()` within `test_invalid_URL_throws_error` -- but we can't, because  we've already called the patcher's `start` method. So then maybe we just define the patch in the `setUp`, but actually `start` it within each individual unit test, but I'm not sure that the extra work of defining it in every method would make sense, given that we'd only be able to consolidate 1 additional test into our general class.

Then with `NdtHtml5SeleniumDriverCustomClassTest` we have that explicit class creation in order to be able to individually test result values. So we could make it a part of `NdtHtml5SeleniumDriverGeneralTest`, but then wed have to define the `mock_driver` within each individual unit test, so as to not override the explicit class creation in the tests in `NdtHtml5SeleniumDriverCustomClassTest`. Let me know if you think there is a clearer solution, thanks.

---

*[tests/test_client_wrapper.py, line 107 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDc_skw4PfVJKccz2YY:-KDd20_32rA5855m49q0:862311698) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/tests/test_client_wrapper.py#L107)):*
Sounds good, I went ahead and added a new test for it.

---

*[tests/test_client_wrapper.py, line 175 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcajFRgUhpNES5Rxq9:-KDct--J-nDgx5ZOONmH:-1962606206) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/tests/test_client_wrapper.py#L175)):*
Sounds good, fixed.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

mtlynch:
Review status: 5 of 7 files reviewed at latest revision, 7 unresolved discussions.

---

*[client_wrapper/client_wrapper.py, line 111 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcTYPFXFB7u-YmoPtf:-KDdRmtXm_VR2TRfUlcY:604269329) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L111)):*
Yeah, that's right. So I see `_click_start_button` as sort of like this:

```
def _click_start_button(driver, errors):
    """Clicks start test button and records start time.
 
    Clicks the start test button for an NDT test and records the start time in
    an NdtResult instance.
 
    Args:
        driver: An instance of a Selenium webdriver browser class.
        errors: A list of errors to append to if errors are encountered in clicking 
           the start button
 
    Returns:
        True if the driver successfully found and clicked the start button.
    """
    try:
        driver.find_element_by_id('websocketButton').click()
        start_button = driver.find_elements_by_xpath(
            "//*[contains(text(), 'Start Test')]")[0]
    except CouldntFindItException: # replace with what Selenium actually returns
        errors.append(TestError( 
            datetime.datetime.now(pytz.utc), 'failed to locate "start test" button'))
        return False
 
    start_button.click()
    return True
```

Does that make sense?

---

*[client_wrapper/client_wrapper.py, line 236 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcZtBV4Yn8cmy2LoPZ:-KDdP4SMckqlFIZYmKiI:-1857764066) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L236)):*
Sorry, I meant to say "I prefer to **avoid** abbrevations" but you got it.

---

*[tests/test_client_wrapper.py, line 22 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6l0lu58WcmxFFgeY:-KDdo4lrOiHtpxx56GtS:-1233931915) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L22)):*
I think we can make this work cleanly. I did a PoC commit, but I may be overlooking complexities in the other test cases. Please take a look and let me know if you think this is generalizable:

https://github.com/rowanv/ndt-e2e-clientworker/compare/ui_automation...mtlynch:test-poc?expand=1

---

*[tests/test_client_wrapper.py, line 109 \[r4\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDdQGB3ZFqjikmiljNu:-KDdQGB3ZFqjikmiljNv:1590259442) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/028f3998329dc4cd1a999469b0fa5bc0207d549b/tests/test_client_wrapper.py#L109)):*
Great test! Thanks

---

*[tests/test_client_wrapper.py, line 121 \[r4\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDdPYlkwoALXCDTi0Zr:-KDdPYlkwoALXCDTi0Zs:931192602) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/028f3998329dc4cd1a999469b0fa5bc0207d549b/tests/test_client_wrapper.py#L121)):*
We can omit the trailing comma

---

*[tests/test_client_wrapper.py, line 123 \[r4\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDdPzIarehYaV9wR9ZX:-KDdPzIarehYaV9wR9ZY:1054682039) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/028f3998329dc4cd1a999469b0fa5bc0207d549b/tests/test_client_wrapper.py#L123)):*
I think we can just use `dates` directly:

>If side_effect is an iterable then each call to the mock will return the next value from the iterable. If any of the members of the iterable are exceptions they will be raised instead of returned.

http://www.voidspace.org.uk/python/mock/mock.html

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

rowanv:
Review status: 5 of 7 files reviewed at latest revision, 6 unresolved discussions.

---

*[client_wrapper/client_wrapper.py, line 111 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDcTYPFXFB7u-YmoPtf:-KDhMrqLEYwqlfDUNMoP:257783731) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/ac9046cbefc1bbe78a35a0635465c88c70f9cb34/client_wrapper/client_wrapper.py#L111)):*
That sounds good. I went ahead and changed `perform_test` so that everything occurs at a similar level of abstraction.

---

*[tests/test_client_wrapper.py, line 22 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6l0lu58WcmxFFgeY:-KDhNAoGx0_e_usS2WTI:-1244209457) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L22)):*
Thanks, that seems like a great solution. Though I'm not sure I see how to generalize it to the remaining 2 test cases that are within their own class, without affecting readability and potentially needing to rework those tests if there was ever a change in the original `client_wrapper.py`, as we'd chatted about in tests/test_client_wrapper.py, line 79 [r2] (raw file). Perhaps it's all right to leave them in their own class since at least 3 more test cases are going to go into that class when each of the metric are tested individually?

---

*[tests/test_client_wrapper.py, line 121 \[r4\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDdPYlkwoALXCDTi0Zr:-KDhHZHvmBHsLmf3cPCe:-2003896356) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/028f3998329dc4cd1a999469b0fa5bc0207d549b/tests/test_client_wrapper.py#L121)):*
Thanks, fixed.

---

*[tests/test_client_wrapper.py, line 123 \[r4\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDdPzIarehYaV9wR9ZX:-KDhI-jCm9HvX4BSbCVH:595388793) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/028f3998329dc4cd1a999469b0fa5bc0207d549b/tests/test_client_wrapper.py#L123)):*
Ah, good catch, thanks.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

mtlynch:
Review status: 5 of 7 files reviewed at latest revision, 6 unresolved discussions.

---

*[client_wrapper/client_wrapper.py, line 103 \[r5\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDhTFlOdsk5lqcQAM0s:-KDhTFlP0_G1pZ6dqXz0:170819623) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/bf59726df74356e8e87699243fff80cd3534b425/client_wrapper/client_wrapper.py#L103)):*
`result` gets passed by reference so the sub-functions don't have to return it back to the caller again. Caller still has it.

This can be simplified to:

```
if not _load_url(driver, url, result):
  return result
```

(ditto for other calls in this function)

---

*[client_wrapper/client_wrapper.py, line 138 \[r5\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDhSmUmhhSPGrglc2HG:-KDhSmUmhhSPGrglc2HH:201938984) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/bf59726df74356e8e87699243fff80cd3534b425/client_wrapper/client_wrapper.py#L138)):*
Should be double quote characters

---

*[tests/test_client_wrapper.py, line 22 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDZ6l0lu58WcmxFFgeY:-KDhSuV9r3QaEDoGimA2:33597615) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/a8a1d15a2c3986b37383a4eda0a462aa1c799373/tests/test_client_wrapper.py#L22)):*
Yeah that's fine.

---

*[tests/test_client_wrapper.py, line 44 \[r5\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDhSx7zcGlo3L2ICzuP:-KDhSx7zcGlo3L2ICzuQ:1337594546) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/bf59726df74356e8e87699243fff80cd3534b425/tests/test_client_wrapper.py#L44)):*
Extraneous whitespace

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

rowanv:
Review status: 5 of 7 files reviewed at latest revision, 5 unresolved discussions.

---

*[client_wrapper/client_wrapper.py, line 103 \[r5\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDhTFlOdsk5lqcQAM0s:-KDhWzdoPjJfoMqACXot:229276453) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/bf59726df74356e8e87699243fff80cd3534b425/client_wrapper/client_wrapper.py#L103)):*
Oh right, thanks, fixed.

---

*[client_wrapper/client_wrapper.py, line 138 \[r5\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDhSmUmhhSPGrglc2HG:-KDhX8cg_nPIgQfC-nXW:-2003896356) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/bf59726df74356e8e87699243fff80cd3534b425/client_wrapper/client_wrapper.py#L138)):*
Thanks, fixed.

---

*[tests/test_client_wrapper.py, line 44 \[r5\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDhSx7zcGlo3L2ICzuP:-KDhXIoid7IlqUS53yXK:-2003896356) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/bf59726df74356e8e87699243fff80cd3534b425/tests/test_client_wrapper.py#L44)):*
Thanks, fixed.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2)*

mtlynch:
:lgtm:

---

Review status: 5 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[client_wrapper/client_wrapper.py, line 143 \[r6\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-KDhZOgYz3WOCQ43jwuc:-KDhZOgZ-Z6s3aQRWf_J:-1296829509) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/112eb2645b25fdfb07414d49ece47b26dcf9a89d/client_wrapper/client_wrapper.py#L143)):*
(minor) We can omit docstring comments like `False if otherwise` because it sort of goes without saying.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/2#-:-KDhZel_5bWmxZWbjNw2:1417088721)*

PR #3, opened 2016-03-25
Breaking up client_wrapper.py into smaller files

mtlynch:
Splitting up client_wrapper.py into smaller files so that it's easier for multiple developers to work on different files in this repo in parallel. This is a straight move with no modifications except for required changes to imports.


nkinkade:
LGTM

PR #5, opened 2016-03-25
Adding a pre-commit hook for client_wrapper

mtlynch:
This adds a build script to client_wrapper that is shared between the
Travis config and the git pre-commit hook.

Also fixes a minor issue in requirements where pytz was in test_requirements
even though it's required for normal application function.


.travis.yml
```diff
@@ -12,12 +12,7 @@ install:
     - pip install yapf==0.6.2
     - pip install pyflakes==1.0.0
 script:
-    # Run unit tests and calculate code coverage.
-    - coverage run --source client_wrapper -m unittest discover
-    # Check that source has correct formatting.
-    - yapf --diff --recursive --style google ./
-    # Run static analysis for Python bugs/cruft.
-    - pyflakes client_wrapper/*.py tests/*.py
+    ./build
```
nkinkade: For clarity, I would rename ./build to ./build.sh.  Extensions aren't needed, of course, but are helpful when glancing at a file name without actually looking in it.

build
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
nkinkade: On pretty much any Linux system you'll find bash at /bin/bash. If you want to be 100% we're running the first bash found in the user's $PATH, rather than making an assumption of the path, then this might be better:

```#!/usr/bin/env bash```

File: build [hunk 3]
```diff
@@ -0,0 +1,13 @@
...
+
+( # Run unit tests and calculate code coverage.
+  coverage run --source client_wrapper -m unittest discover
+) &&
+( # Check that source has correct formatting.
+  yapf --diff --recursive --style google ./
+) &&
+( # Run static analysis for Python bugs/cruft.
+  pyflakes client_wrapper/*.py tests/*.py
+)
```
nkinkade: Since these commands are all running sequentially, is there a reason to have them running in subshells (by wrapping them parens)?  I think you can get away with simply this:

```
coverage run --source client_wrapper -m unittest discover &&
yapf --diff --recursive --style google ./ &&
pyflakes client_wrapper/*.py tests/*.py
```

hooks/pre-commit
```diff
@@ -0,0 +1 @@
+./build
```
nkinkade: If you decide to change the name of the build script from _build_ to _build.sh_, don't forget to change this.

.travis.yml
mtlynch: Yeah, it looks weird to me too, but style guide discourages .sh

>Executables should have no extension (strongly preferred) or a .sh extension. Libraries must have a .sh extension and should not be executable.
It is not necessary to know what language a program is written in when executing it and shell doesn't require an extension so we prefer not to use one for executables.

https://google.github.io/styleguide/shell.xml?showone=File_Extensions#File_Extensions

hooks/pre-commit
mtlynch: ack

build
mtlynch: Is there a way to do this but keep the comments around? That's what tripped me up.

I tried this:
```
# Run unit tests and calculate code coverage.
coverage run --source client_wrapper -m unittest discover && \
 \ # Check that source has correct formatting.
yapf --diff --recursive --style google ./ && \
 \ # Run static analysis for Python bugs/cruft.
pyflakes client_wrapper/*.py tests/*.py
```

But then it fails:

```
$ ./build 
........
----------------------------------------------------------------------
Ran 8 tests in 0.489s

OK
./build: line 7:  #: command not found
./build: line 9:  #: command not found
```

build
mtlynch: Whoops, I don't think my comment went through. The style guide requires /bin/bash:

>Executables must start with #!/bin/bash

https://google.github.io/styleguide/shell.xml?showone=Which_Shell_to_Use#Which_Shell_to_Use

.travis.yml
nkinkade: Okay, cool.

build
nkinkade: If the style guide says so, but this script would fail on, for example, FreeBSD, where bash is found at /usr/local/bin/bash.  I suspect the intent of the the line "Executables must start with #!/bin/bash" is less about absolutely using that exact string of characters than in being specific about the shell the script should be using.  Anyway....

build
nkinkade: Yes, just get rid of the line continuation characters and it should work:

```
# Run unit tests and calculate code coverage.
coverage run --source client_wrapper -m unittest discover &&
# Check that source has correct formatting.
yapf --diff --recursive --style google ./ &&
# Run static analysis for Python bugs/cruft.
pyflakes client_wrapper/*.py tests/*.py
```

build
mtlynch: I'd like to leave in for now just for the sake of consistency. It's unlikely to run on any system with an unusual /bin/bash location.

build
mtlynch: Fixed thanks!

mtlynch:
LGTM from @nkinkade

PR #8, opened 2016-03-29
Refactor name constants into a specific name module

mtlynch:
This change pulls all the name constants (for NDT clients, browsers, OSes)
into a separate file of just constants. This eliminates string duplication
and reduces the risk of string mismatches (e.g. 'ndtjs' vs 'ndt_js').

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/8)

### DIFF
--- a/client_wrapper/client_wrapper.py
+++ b/client_wrapper/client_wrapper.py
@@ -15,12 +15,11 @@
 import argparse
 
 import html5_driver
-
-NDT_HTML5_CLIENT = 'ndt_js'
+import names
 
 
 def main(args):
-    if args.client == NDT_HTML5_CLIENT:
+    if args.client == names.NDT_HTML5:
         driver = html5_driver.NdtHtml5SeleniumDriver(args.browser,
                                                      args.client_url,
                                                      timeout=20)
@@ -46,7 +45,7 @@ if __name__ == '__main__':
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     parser.add_argument('--client',
                         help='NDT client implementation to run',
-                        choices=(NDT_HTML5_CLIENT,),
+                        choices=(names.NDT_HTML5,),
                         required=True)
     parser.add_argument('--browser',
                         help='Browser to run under (for browser-based client)',
--- a/client_wrapper/html5_driver.py
+++ b/client_wrapper/html5_driver.py
@@ -21,6 +21,7 @@ from selenium.webdriver.support import ui
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.common import exceptions
 
+import names
 import results
 
 
@@ -78,9 +79,9 @@ def _create_browser(browser):
         An instance of a Selenium webdriver browser class corresponding to
         the specified browser.
     """
-    if browser == 'firefox':
+    if browser == names.FIREFOX:
         return webdriver.Firefox()
-    if browser in ['chrome', 'edge', 'safari']:
+    if browser in [names.CHROME, names.EDGE, names.SAFARI]:
         raise NotImplementedError
     raise ValueError('Invalid browser specified: %s' % browser)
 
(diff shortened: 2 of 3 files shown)


mtlynch:
>naming the new module something like "constants.py" instead of "names.py"

Yeah, this is a reaction from mlab-ns which has names like "util" and "constants." It seems like what ended up happening there is that name is so generic that it becomes unclear what's supposed to be there, so it ends up getting filled with things that could have narrower scope.

For example:

https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/util/constants.py

I agree that "names" feels a bit unexpected, but I'm having trouble coming up with something that's clear, but also doesn't become a dumping ground. I considered `browser_names.py`, `os_names.py`, `client_names.py` but that feels like too much clutter.

Any thoughts?

nkinkade:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[client_wrapper/client_wrapper.py, line 18 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/8#-KE3Wd3Gv9kDxHEm-q-8:-KE3Wd3Gv9kDxHEm-q-9:1220520173) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/d5554f561c3da9fbe83300bf1fe474cbf2eab16d/client_wrapper/client_wrapper.py#L18)):*
Would a name like "constants" be more intuitive than "names"?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/8)*

nkinkade:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[client_wrapper/client_wrapper.py, line 18 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/8#-KE3Wd3Gv9kDxHEm-q-8:-KE3iele0dVfJH0XCulZ:-2045745786) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/d5554f561c3da9fbe83300bf1fe474cbf2eab16d/client_wrapper/client_wrapper.py#L18)):*
I made this comment, but forgot to publish it, so this is showing up after the fact.  

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/8)*

nkinkade:
Sorry.  Not the first time I've done this, but I had made a draft comment here in Reviewable, but forgot to publish it.

Out of curiosity, is there ever going to be a case where where we name Firefox something other than "firefox", or Chrome something other than "chrome", etc.? I can see NDT_HTML5 changing, but those others seem like names that will never likely change, especially when the constant name is identical, except all caps.   This makes me wonder whether we're adding abstraction here for things that may not need abstraction.

In any case, I can't think of anything better than "names" that isn't equally as generic, so if this is the thing, then I suppose "names" is fine.

LGTM.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/8#-:-KE3j2TpAnQGsjBnDydr:-28475292)*

mtlynch:
> I can see NDT_HTML5 changing, but those others seem like names that will never likely change, especially when the constant name is identical, except all caps. This makes me wonder whether we're adding abstraction here for things that may not need abstraction.

I agree with you and it doesn't feel 100% right to me either, but I think there's value in making these into defined constants. The biggest is that it lets you easily grep for references to "Firefox the symbolic name" as opposed to any instance of the word "Firefox" (e.g. in comments). It also makes it easy to change the canonical name in a single place if we decide that we decide to go from like "ie" (not yet supported) to "msie."

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/8#-:-KE7O1WZ_FFC-wGhfCwO:1413865191)*

PR #17, opened 2016-04-05
Excluding selenium install script from code coverage settings

mtlynch:
This is an install script for dependencies and we don't really want it
considered under code coverage.


fernandalavalle:
LGTM

PR #26, opened 2016-04-12
LocalServer Class and tests

fernandalavalle:
Added a class for creating and starting a local server for hosting local clients.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/26)


mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[client_wrapper/local_server.py, line 6 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAVa1S-rS4_PNIkcBr:-KFAVa1S-rS4_PNIkcBs:-1688286142) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/client_wrapper/local_server.py#L6)):*
`LocalServer` is a bit generic. `NdtClientHTTPServer`? Web or HTTP should be in there someplace.

---

*[client_wrapper/local_server.py, line 7 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAVqAkBet2DLxgiSmx:-KFAVqAkBet2DLxgiSmy:-1947596401) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/client_wrapper/local_server.py#L7)):*
We should add a class docstring that explains what this is and documents the attributes.

---

*[client_wrapper/local_server.py, line 8 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAVVDql9re6gW9kstC:-KFAVVDql9re6gW9kstD:-153330999) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/client_wrapper/local_server.py#L8)):*
Caller should point it to the local directory it's going to serve.

---

*[client_wrapper/local_server.py, line 13 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAVxAIbm2TXs1UeXN8:-KFAVxAIbm2TXs1UeXN9:1374678760) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/client_wrapper/local_server.py#L13)):*
Suggest "bound to a random available port."

---

*[client_wrapper/local_server.py, line 29 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAV3tsFGb_ZtE0sCdg:-KFAV3tsFGb_ZtE0sCdh:1197281875) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/client_wrapper/local_server.py#L29)):*
Code generally drops the 'h' when abbreviating asynchronous, so just `async` is standard.

---

*[tests/test_local_server.py, line 15 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAVE6xW5AJVEk3Y4dd:-KFAVE6xW5AJVEk3Y4de:678904379) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/tests/test_local_server.py#L15)):*
`LocalServerTest`

---

*[tests/test_local_server.py, line 41 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAXWLh34GgHMggd9X9:-KFAXWLiG_JSB58ygy1c:392207777) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/tests/test_local_server.py#L41)):*
Unit tests shouldn't call into private methods of the class they're testing. If we find it's hard to test it otherwise, we might want to split this into two classes. One that's synchronous and another that's a wrapper around that class that allows it to start asynchronously. Then some of the private methods become public when they're in a separate class and we can mock out the synchronous version when we're testing the async version.

---

*[tests/test_local_server.py, line 57 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAYC35iR-9bn2cO7kM:-KFAYC35iR-9bn2cO7kN:584959704) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/tests/test_local_server.py#L57)):*
We should assert that `serve_forever` was called on our mock object.

---

*[tests/test_local_server.py, line 58 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFAY-drXOoGcB0CPuZ6:-KFAY-drXOoGcB0CPuZ7:1243091481) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/1e026a3296253077565444283f626b9a0b49d627/tests/test_local_server.py#L58)):*
We should assert that `shutdown` was called on our mock object.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26)*

mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 17 unresolved discussions.

---

*[client_wrapper/ndt_client_http_server.py, line 16 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFF8ls6abQFNep3QxgO:-KFF8ls6abQFNep3QxgP:-690245431) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L16)):*
This is pretty odd. Why hide a class within a function? If we don't want clients outside the module to instantiate it, can we make it a private class (prefix a `_` to the name)

This looks close to the factory design pattern, so if that's what we want, we should make it look like the standard pattern:

https://en.wikipedia.org/wiki/Factory_method_pattern

If you have access to O'Reilly eBooks (either through their site or through Safari Books Online), *Head First Design Patterns* has a good explanation of the Factory pattern:

http://shop.oreilly.com/product/9780596007126.do

I'm not saying that's necessarily what we want. We may be able to do what we want with just a private class.

---

*[client_wrapper/ndt_client_http_server.py, line 26 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFF8yA2jFbcT68lrgCC:-KFF8yA2jFbcT68lrgCD:2021625501) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L26)):*
Isn't this backwards?

---

*[client_wrapper/ndt_client_http_server.py, line 32 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFCMl1MzQDIb9kf_tk:-KFFCMl1MzQDIb9kf_tl:2091162172) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L32)):*
It's a bit unclear what we mean by "represents a SimpleHTTPServer". It's not a subclass of that class.

Suggest "A simple HTTP server that listens on a randomly assigned port and serves GET requests for files in a specified directory"

---

*[client_wrapper/ndt_client_http_server.py, line 36 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFDN35l8nREPgWyTyY:-KFFDN35l8nREPgWyTyZ:593277767) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L36)):*
Suggest keeping the order of `port`, `client_path` consistent throughout

---

*[client_wrapper/ndt_client_http_server.py, line 40 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFDKHMwDiKwws7wOAo:-KFFDKHMwDiKwws7wOAp:-2078781925) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L40)):*
Suggest making this read-only since we don't really support letting the caller change the path after instantiation.

---

*[client_wrapper/ndt_client_http_server.py, line 69 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFCnHYP1ybw96D_LrK:-KFFCnHZm8BVwxqZ88vJ:-1341382708) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L69)):*
I think this can reduce to 

```
threading.Thread(target=self._tcp_server.serve_forever).start()
```

---

*[tests/test_ndt_client_http_server.py, line 42 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFGPeyPTk2rXqXXCcV:-KFFGPezYnFRMmTmAqo8:-230700095) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/tests/test_ndt_client_http_server.py#L42)):*
I don't really understand the difference between this test and the final test. This one just seems like a superset of the other.

---

*[tests/test_ndt_client_http_server.py, line 60 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFEPbSdTamSy8inpvf:-KFFEPbSdTamSy8inpvg:-667041855) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/tests/test_ndt_client_http_server.py#L60)):*
We can cut out `assures_that` since that's kind of implied, also we can drop the `h` in asynch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26)*

fernandalavalle:
Review status: 0 of 2 files reviewed at latest revision, 17 unresolved discussions.

---

*[client_wrapper/ndt_client_http_server.py, line 16 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFF8ls6abQFNep3QxgO:-KFFGqOmmVjASJJSN4Xm:-1129970877) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L16)):*
Because we pass in a class not an instance to the server. Wasn't sure how else to dynamically assign an instance variable without actually creating an object and this didn't seem too crazy. Reading through the Wikipedia article you linked me to, doesn't seem exactly like what we want?  But I've been wrong before.

---

*[client_wrapper/ndt_client_http_server.py, line 26 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFF8yA2jFbcT68lrgCC:-KFFH3pGEcTi6K_6EmpA:-1903518047) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L26)):*
I don't think so? It works, and changing the order gives a path that's just '\' (or path). 

---

*[client_wrapper/ndt_client_http_server.py, line 40 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFDKHMwDiKwws7wOAo:-KFFJOmb5eLpHP-3xARn:-1557018268) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L40)):*
We also don't really need them to read it after, right? Should we just make it private?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26)*

fernandalavalle:
Review status: 0 of 2 files reviewed at latest revision, 17 unresolved discussions.

---

*[client_wrapper/ndt_client_http_server.py, line 26 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFF8yA2jFbcT68lrgCC:-KFFNMcJdu2BzhQYpiQg:740524452) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L26)):*
hmm actually this originally shows files in self._root, but you can't navigate down the tree. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26)*

mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 18 unresolved discussions.

---

*[client_wrapper/ndt_client_http_server.py, line 16 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFF8ls6abQFNep3QxgO:-KFFN4q8ozXeb4mSn0di:-1620577816) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L16)):*
Ohhhhhh so any parameterized behavior has to be baked into the class itself rather than instance variables. Okay, that's super confusing, but I think that is kind of what Python is forcing us to do because we need to pass in a class instead of an instance.

The implementation is fine, but we need to document a lot more to make it super obvious what we're doing because this is really hard to understand. I kept trying to figure out who was calling the constructor we were defining. One thing to start, suggest changing the function name to explicitly say class like `create_custom_http_handler_class`.

---

*[client_wrapper/ndt_client_http_server.py, line 26 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFF8yA2jFbcT68lrgCC:-KFFKMRK7jiizRlPE4Uy:-1955888197) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L26)):*
I thought the root is something like `/home/fernanda/www` and path is something like `ndt/run_client.html`, so 

```
os.path.join('ndt/run_client.html', '/home/fernanda/www')
>>>  '/home/fernanda/www'
```
whereas the opposite:

```
os.path.join('/home/fernanda/www', 'ndt/run_client.html')
>>>  '/home/fernanda/www/ndt/run_client.html'
```

is what I'd expect

---

*[client_wrapper/ndt_client_http_server.py, line 40 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFDKHMwDiKwws7wOAo:-KFFNgQ-VGoWPWbN5dKh:-499175790) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L40)):*
Oh yeah, private's even better.

---

*[client_wrapper/ndt_client_http_server.py, line 54 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFOjkdJ_ylhsTRM3px:-KFFOjkdJ_ylhsTRM3py:-150991256) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/client_wrapper/ndt_client_http_server.py#L54)):*
Would we want `BaseHTTPServer.HTTPServer` instead?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26)*

fernandalavalle:
Review status: 0 of 2 files reviewed at latest revision, 17 unresolved discussions.

---

*[tests/test_ndt_client_http_server.py, line 42 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFGPeyPTk2rXqXXCcV:-KFG-lSFEQ8_MLu8qWam:-1916743014) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/tests/test_ndt_client_http_server.py#L42)):*
Just the order. The last one checks port while the server is still running to prove that its still responsive while its serving.  You think its redundant/unnecessary?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26)*

mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 18 unresolved discussions.

---

*[client_wrapper/ndt_client_http_server.py, line 7 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFKVsxILTS53Hiygw0l:-KFKVsxILTS53Hiygw0m:-1188404936) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/3ba507a0bc730e628109b31bdb109305ee43908d/client_wrapper/ndt_client_http_server.py#L7)):*
We should *start* with changing the function name, but we need more documentation. 

We normally shouldn't assume the usage of a function, but in this case I feel like it would be so strange to a reader to see us creating a custom class in a method that we should say specifically that we need to do it to create a class to give to `BaseHTTPServer.HTTPServer`.

---

*[client_wrapper/ndt_client_http_server.py, line 27 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFKQ8fWkEdAYN3cL7Lh:-KFKQ8fWkEdAYN3cL7Li:1235021834) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/3ba507a0bc730e628109b31bdb109305ee43908d/client_wrapper/ndt_client_http_server.py#L27)):*
We're really just trying to remove the leading slash in a URL, right? I think we should go really simple and say

```
# Strip leading slash from URL
if path and path[0] == '/':
  path = [1:]
```

Because this is a URL path, not really an OS path. On Windows if we do `os.path.relpath('/foo/bar.txt', '/')` it returns `foo\bar.txt` which is a bit unexpected because I expect this to still be a URL path and now it's a file path. The function will still work because os.path.join is pretty liberal with what it takes in, but it's confusing that os.path.join is joining either a URL path or a file path depending on if the `path` parameter had a leading slash.

---

*[client_wrapper/ndt_client_http_server.py, line 57 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFKSV57gAbfJJsktpKl:-KFKSV57gAbfJJsktpKm:-290901836) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/3ba507a0bc730e628109b31bdb109305ee43908d/client_wrapper/ndt_client_http_server.py#L57)):*
Single space after `#`
https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

---

*[tests/test_ndt_client_http_server.py, line 42 \[r2\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFFGPeyPTk2rXqXXCcV:-KFKUUqGCus9HI784ZnR:-783537315) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/f490062f25673d5dbe0a46258caf5aaaed3141c9/tests/test_ndt_client_http_server.py#L42)):*
Yeah, I think it's too confusing. Checking ports after we've stopped listening doesn't really make any sense. We should keep the first test but move the port check to just after async_start.

Also suggest doing all the post-start assertions after calling async_start, then doing all the post-stop assertions after calling `stop`, so like:

```
self.server.async_start()
self.assertEqual(8888, self.server.port)
self.assertTrue(self.mock_http_server.serve_forever.called)
        ndt_client_http_server.BaseHTTPServer.HTTPServer.assert_called_with(
            ('', 0), self.mock_handler)
self.assertFalse(self.mock_http_server.shutdown.called)
        
self.server.stop()
self.assertTrue(self.mock_http_server.shutdown.called)
```

---

*[tests/test_ndt_client_http_server.py, line 21 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFKWKwnJVftFnQ-VqgM:-KFKWKwoARproCo2k81v:1440262316) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/3ba507a0bc730e628109b31bdb109305ee43908d/tests/test_ndt_client_http_server.py#L21)):*
do we actually need to name them?

---

*[tests/test_ndt_client_http_server.py, line 27 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFKTbWyUETfwqT8IdZj:-KFKTbWyUETfwqT8IdZk:-1364856949) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/3ba507a0bc730e628109b31bdb109305ee43908d/tests/test_ndt_client_http_server.py#L27)):*
Isn't this redundant because it's also happening on the last line? I prefer the last line version because I think it's more obvious what's getting mocked.

---

*[tests/test_ndt_client_http_server.py, line 39 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFKTls6xVITJBRIBesX:-KFKTls6xVITJBRIBesY:1196752945) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/3ba507a0bc730e628109b31bdb109305ee43908d/tests/test_ndt_client_http_server.py#L39)):*
I recommend avoiding defining things in `setUp` unless they apply to all of the tests. Splitting up the logic can clean up the individual test logic, but it comes at the cost of making the tests less clear. So someone reading `test_ndt_client_http_server_async_start_starts_and_returns_successfully` might not immediately understand why the port should be `8888`, so they'd have to go read the `setUp`.

---

*[tests/test_ndt_client_http_server.py, line 40 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFKTWU2a6XwWZWTMeiM:-KFKTWU3U8WcPr1mzku5:-491519861) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/3ba507a0bc730e628109b31bdb109305ee43908d/tests/test_ndt_client_http_server.py#L40)):*
> 80 characters

---

*[tests/test_ndt_client_http_server.py, line 47 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFKUdKvIUSAz08rv_IK:-KFKUdKvIUSAz08rv_IL:-829781624) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/3ba507a0bc730e628109b31bdb109305ee43908d/tests/test_ndt_client_http_server.py#L47)):*
We shouldn't be peeking into private variables in tests. It seems like the following line makes this redundant anyway because if it's launching the server, then we're good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26)*

mtlynch:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[client_wrapper/ndt_client_http_server.py, line 13 \[r4\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFP_NnfwJ3clOfoFzz8:-KFP_NnggmGVo60MCfEn:-2126995363) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/176125e93357fb44a6695cbd2db02d574dbdf492/client_wrapper/ndt_client_http_server.py#L13)):*
This is a bit hard to read. Two things I've been taught about clear technical writing I think apply here:

* It's hard for humans to parse technical language when the author writes in the passive voice. The easiest sentence structure to parse is [subject] [verb] [direct object], whereas passive voice creates more complex structures. Dropping subjects introduces ambiguity in the sentence.
* Avoid "nounifying" verbs. Verbs drive the sentence forward in a clear way. So "class creation takes place within a function" is harder to parse than "the function creates the class."

So again, sometimes we have to write code that's unintuitive, but then it's on us to give really clear explanations of what's going on. The code is unavoidably confusing, so the bar for clarity in this docstring is high.

Also, the first sentence seems to be a sentence fragment (I think we're missing a leading "The" or something).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-:-KFPbcJx8aWh4Q0l4R6H:1417088721)*

mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions.

---

*[client_wrapper/ndt_client_http_server.py, line 37 \[r5\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26#-KFoaVBTVOaFcyKKOYL-:-KFoaVBUZXhad0psqyLs:1278331507) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/82a4c0bde99bdd0d2712dfb1749c741f85de3262/client_wrapper/ndt_client_http_server.py#L37)):*
Can we rename the file to `http_server.py` and the class to `StaticFileHttpServer`? I initially tried to name my server consistently with yours and I'm seeing that to instantiate it, I have to do this super long string `ndt_client_replay_http_server.NdtClientReplayHTTPServerTest(...)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/26)*

PR #32, opened 2016-04-13
Fixing time logging in HTML5 driver

mtlynch:
This makes a few changes to recording timestamps during the test that are a
bit intermixed in the code:

1. Changes recording of start time and end time so that they record when the
underlying client process began and ended, not when the the actual tests began
(the revised behavior matches the spec).

2. Changes recording of s2c/c2s end_time values so that it actually waits for the
element it's waiting for before recording the time (it would previously record
the time without waiting, which was a bug and caused it to record incorrect
times).

In addition, refactors html_driver a bit so that there is less mixing of
abstraction layers within functions.

Lastly, deletes the freezegun unit test as it was not providing additional
coverage over the final timing test and it's an additional unneeded
dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/32)

### DIFF
--- a/client_wrapper/html5_driver.py
+++ b/client_wrapper/html5_driver.py
@@ -19,11 +19,17 @@ import datetime
 import pytz
 from selenium import webdriver
 from selenium.webdriver.support import ui
-from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support import expected_conditions
 from selenium.common import exceptions
 
 import names
 import results
+
+# TODO(mtlynch): Define all error strings as public constants so we're not
+# duplicating strings between production code and unit test code.
+ERROR_C2S_NEVER_STARTED = 'Timed out waiting for c2s test to begin.'
+ERROR_S2C_NEVER_STARTED = 'Timed out waiting for c2s test to begin.'
+ERROR_S2C_NEVER_ENDED = 'Timed out waiting for c2s test to end.'
 
 
 class NdtHtml5SeleniumDriver(object):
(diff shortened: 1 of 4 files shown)


pboothe:
:lgtm:

---

Reviewed 2 of 4 files at r1.
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[client_wrapper/html5_driver.py, line 116 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/32#-KFKi6f-aN8LR81_ooIK:-KFKi6f-aN8LR81_ooIL:-1959997418) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/7b9d10e2be2fc02d723e200277aa84fd89562918/client_wrapper/html5_driver.py#L116)):*
Nit: "any" -> "an" or "a specified"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/32#-:-KFKi4A8UgItOzmEUB6K:1417088721)*

mtlynch:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[client_wrapper/html5_driver.py, line 116 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/32#-KFKi6f-aN8LR81_ooIK:-KFKmajaMvAysqKZ0pG6:1154580994) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/7b9d10e2be2fc02d723e200277aa84fd89562918/client_wrapper/html5_driver.py#L116)):*
Fixed, thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/32)*

PR #35, opened 2016-04-20
Adding CORS support to fake mlab-ns server

mtlynch:
CORS support is needed if, for example, we have a web server on
localhost:8123 that wants to refer the client to a fake mlab-ns server on
localhost:8321. Without a CORS header, this is considered a violation of
same origin policy and the browser would block the response.

Adding CORS better simulates real mlab-ns.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/35)

### DIFF
--- a/client_wrapper/fake_mlabns.py
+++ b/client_wrapper/fake_mlabns.py
@@ -60,6 +60,9 @@ class _FakeMLabNsHandler(BaseHTTPServer.BaseHTTPRequestHandler):
         """
         self.send_response(200)
         self.send_header('Content-type', 'application/json')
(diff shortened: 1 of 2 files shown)


fernandalavalle:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/35#-:-KFtNeUz6ex5zbSID-Oe:1417088721)*

PR #38, opened 2016-04-25
Pulling common browser functions to common browser module

mtlynch:
Pulling out functions from the HTML5 driver that are reusable in other
browser-based NDT tests and putting them in a common browser function
module.


mtlynch:
PR is too big, breaking it up

PR #40, opened 2016-04-25
Creates StaticFileHTTPServer

fernandalavalle:
This PR adds a class for serving static local files that can be used to test locally hosted clients.


mtlynch:
Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[client_wrapper/http_server.py, line 136 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGHyq1ACyiI6KW3Ids_:-KGHyq1ACyiI6KW3Idsa:-660775608) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/65d930faf5682ed09b517fda6bd0c5771496527c/client_wrapper/http_server.py#L136)):*
Suggest making this private since we're only unit testing `StaticFileHTTPServer`

---

*[client_wrapper/http_server.py, line 140 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGHxwaap0JUXK8GJFGD:-KGHxwaap0JUXK8GJFGE:890754113) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/65d930faf5682ed09b517fda6bd0c5771496527c/client_wrapper/http_server.py#L140)):*
Can we make this a read-only property?

---

*[client_wrapper/http_server.py, line 166 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGHyCG9Tt0by1oTkKNZ:-KGHyCG9Tt0by1oTkKN_:1212793376) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/65d930faf5682ed09b517fda6bd0c5771496527c/client_wrapper/http_server.py#L166)):*
Think this is obsolete

---

*[client_wrapper/http_server.py, line 185 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGHyMIldcXF9cPpyF7J:-KGHyMIldcXF9cPpyF7K:1913952949) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/65d930faf5682ed09b517fda6bd0c5771496527c/client_wrapper/http_server.py#L185)):*
I realized that `self.server_address` is the more direct way to this. See:

https://github.com/m-lab/ndt-e2e-clientworker/blob/master/client_wrapper/fake_mlabns.py#L40

---

*[client_wrapper/http_server.py, line 188 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGHyX7DkIMJaCxLY4sJ:-KGHyX7DkIMJaCxLY4sK:-403656332) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/65d930faf5682ed09b517fda6bd0c5771496527c/client_wrapper/http_server.py#L188)):*
Suggest the name `close` because that way it works with `contextlib.closing`, like here:

https://github.com/m-lab/ndt-e2e-clientworker/blob/master/client_wrapper/client_wrapper.py#L33

---

*[tests/test_http_server.py, line 91 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGHzWsJgIYXm2xxKO84:-KGHzWsJgIYXm2xxKO85:-688100578) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/65d930faf5682ed09b517fda6bd0c5771496527c/tests/test_http_server.py#L91)):*
Eh, if we mock this out, we're mocking out about half the code we're trying to test. This is kind of a hairy situation because it's hard to do the mocks right while we inherit from `HTTPServer`.

My suggestion is to mock nothing and just create a file like `tests/testdata/static_file_server/foo.html` and start a real `StaticFileHTTPServer` pointed at that directory and then the unit test is just that it actually serves the right file. In general, I don't like writing unit tests that depend on the filesystem, but this is the best I can think of for this case.

We're doing something similar for `FakeMLabNs` except not hitting the real filesystem:

https://github.com/m-lab/ndt-e2e-clientworker/blob/master/tests/test_fake_mlabns.py

Another way to do this is if we could test `CustomRootHTTPRequestServer` independently and somehow verify that the custom path logic is working, but I'm not sure what APIs are available that will let us do that.

---

*[tests/test_http_server.py, line 120 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGHxkxT94z6kryB88Fj:-KGHxkxT94z6kryB88Fk:1207811914) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/65d930faf5682ed09b517fda6bd0c5771496527c/tests/test_http_server.py#L120)):*
Single quotes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40)*

mtlynch:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions.

---

*[tests/test_http_server.py, line 103 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGVz8rhqk66RRxkN3po:-KGVz8riEE5TsbxX4kWd:2063038480) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/97613245b185badcbe39da145058427551d8be58/tests/test_http_server.py#L103)):*
I'd like to avoid situations where we have to remember to call `close`. Suggestions in descending order of preference, but I'd be fine with all of them:

1. Move `async_start` to the server's constructor so that the caller always calls it within a `with contextlib.closing(...)` block.
1. Add `self.addCleanup(self.server.close)` to `setUp`.
1. Add a `tearDown` method and call `self.server.close()` there.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-:-KH0hzBHkRfG16w0bbw_:1417088721)*

fernandalavalle:
Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[tests/test_http_server.py, line 103 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGVz8rhqk66RRxkN3po:-KHaFusX7MgqcDb8J5_v:-794170359) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/97613245b185badcbe39da145058427551d8be58/tests/test_http_server.py#L103)):*
<details><summary><i>Previously, mtlynch (Michael Lynch) wrote…</i></summary>
> I'd like to avoid situations where we have to remember to call `close`. Suggestions in descending order of preference, but I'd be fine with all of them:
> 
> 1. Move `async_start` to the server's constructor so that the caller always calls it within a `with contextlib.closing(...)` block.
> 1. Add `self.addCleanup(self.server.close)` to `setUp`.
> 1. Add a `tearDown` method and call `self.server.close()` there.</details>
I'm not sure I understand 1. So you propose to start the server as soon as we instantiate it? How do we ensure the caller always calls within `with contextlib.closing(...)` ?

The other two options don't currently work with the second test because an error is raised in  `_create_server()` so close is shutting down a None self._http_server.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40)*

mtlynch:
Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[tests/test_http_server.py, line 103 \[r3\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40#-KGVz8rhqk66RRxkN3po:-KHaUB3GZU7F2K5Zt1ry:1461501785) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/97613245b185badcbe39da145058427551d8be58/tests/test_http_server.py#L103)):*
> ```Python
> 
>         self.assertEqual(actual_response, expected_response)
>         self.server.close()
> ```

>I'm not sure I understand 1. So you propose to start the server as soon as we instantiate it? 

Yeah, it's what I did in [`ReplayHTTPServer`](https://github.com/m-lab/ndt-e2e-clientworker/blob/master/client_wrapper/http_server.py#L66-84). I don't love it because I don't like to do complicated stuff in the constructor, but I couldn't figure out another way to let the caller manage the server using a context handler.

>How do we ensure the caller always calls within with contextlib.closing(...) ?

We can't enforce it through code, but we can document that they should.

>The other two options don't currently work with the second test because an error is raised in _create_server() so close is shutting down a None self._http_server.

This actually usefully reveals a bug in the `close` method then. The block in `close` should be like

```python
if self._http_server:
  self._http_server.shutdown()
```

It seems like more sane behavior that calling `close` on an un-started server is a no-op rather than throwing a null reference exception.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/40)*

Issue #43, opened 2016-04-26
Race condition in ReplayHTTPServer causing flaky tests


mtlynch:
There seems to be a race condition in `ReplayHTTPServer` as unit tests will occasionally break randomly:

```
======================================================================
FAIL: test_creating_server_creates_correct_threads_and_processes (tests.test_http_server.ReplayHTTPServerTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/m-lab/ndt-e2e-clientworker/tests/test_http_server.py", line 61, in test_creating_server_creates_correct_threads_and_processes
    self.assertTrue(self.mock_mlabns_server.serve_forever.called)
AssertionError: False is not true
```

But then re-running Travis fixes it.

mtlynch:
D'oh still seeing failures from this.

PR #44, opened 2016-04-26
Reducing impact of race condition in ReplayHTTPServer

mtlynch:
There is a race condition in ReplayHTTPServer in that one thread needs to
execute serve_forever, which blocks forever, but other threads need to know
that it has started serving.

I wasn't able to figure out a way to eliminate the race condition
completely, but I believe this reduces the impact to the point that it is
fixed for practical purposes. In practice, there will be a delay of several
milliseconds between the client making an HTTP request to the replay server
then making another HTTP request to the fake mlab-ns server.

This is mainly to fix flakiness in unit tests (issue #43), which it appears
to fix.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/44)

### DIFF
--- a/client_wrapper/http_server.py
+++ b/client_wrapper/http_server.py
@@ -81,7 +81,7 @@ class ReplayHTTPServer(object):
         self._start_mitmdump_async()
 
     def _start_fake_mlabns_async(self):
-        """Start the fake mlab-ns server in a background thread
+        """Start the fake mlab-ns server in a background thread.
 
         Start the fake mlab-ns server in a background thread, but block until
         that thread begins.
--- a/client_wrapper/http_server.py
+++ b/client_wrapper/http_server.py
@@ -67,6 +67,7 @@ class ReplayHTTPServer(object):
         self._mlabns_server = mlabns_server
         self._replay_filename = replay_filename
         self._mlabns_thread = None
+        self._mlabns_serving_event = threading.Event()
         self._server_proc = None
         self._start_async()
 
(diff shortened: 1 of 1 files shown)


pboothe:
Makes sense, bummer that the race condition still exists. Does calling `_start_async()` in `__init__()` fall afoul of the "don't do too much work in constructors" rule? Does that rule even exist for Python? It's not in the Python style guide, but it is in the C++ style guide, and I'm never sure how much to import from one to the other. Looking more into it, it looks like it's a side effect of the fact that exceptions are forbidden in C++ (but not in Python) so we're all good here.

LGTM

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/44#-:-KGNb2uj2FazZmEk5wql:-226993375)*

mtlynch:
> Does calling _start_async() in __init__() fall afoul of the "don't do too much work in constructors" rule?

Yeah, I don't like doing that, but I don't know another way of making it work cleanly with a context handler. See:
https://github.com/m-lab/ndt-e2e-clientworker/blob/master/client_wrapper/client_wrapper.py#L33-37

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/44#-:-KGNdTe2iE8wYuW98U3S:-1207562893)*

PR #47, opened 2016-04-27
Adding a partial implementation of the Banjo driver

mtlynch:
Adds a partial selenium driver for the Banjo client. Adds a unit test for
the currently implemented functionality.

Also fixes a minor issue with the TestResult constructor. It turns out that
using the default parameter of [] was causing all instances that used that
constructor to share the same list, which breaks unit tests. This fixes
the constructor and both client drivers' usage of the constructor.

The Banjo driver differs from the HTML5 driver in that the Banjo driver
uses a private _BanjoUiFlowWrapper so that it doesn't keep having to
pass around the "driver" and "result" parameters to different module-
level functions, but perform_test still is idempotent, which was the reason
the HTML5 driver didn't use instance variables to pass around data
in the first place. We should eventually refactor the HTML5 driver to follow
the same pattern.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/47)

### DIFF
--- a/client_wrapper/banjo_driver.py
+++ b/client_wrapper/banjo_driver.py
@@ -1,0 +1,97 @@
+# Copyright 2016 Measurement Lab
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


fernandalavalle:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/47#-:-KGSL527xXhJXOS9qzVM:1417088721)*

PR #50, opened 2016-04-29
Fixes a missing elif

mtlynch:
The if should be an elif, otherwise we always raise a spurious ValueError
when Banjo is the selected client.


nkinkade:
LGTM

PR #55, opened 2016-05-03
Fixing client wrapper so it always quit()s Selenium

mtlynch:
This fixes a bug in that the client_wrapper was just calling close() on
the Selenium driver, which closes the browser window, but does not free
resources associated with the browser driver. To do that, we need to call
quit().

This changes create_browser to a context manager so that callers can call
it in a with block that automatically calls quit() at the end. Updates the
unit tests to match.

Small refactoring of apply_patches_for_create_browser so that it's now shared
between both the banjo tests and the HTML5 tests.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/55)

### DIFF
--- a/client_wrapper/banjo_driver.py
+++ b/client_wrapper/banjo_driver.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from __future__ import division
-import contextlib
 import datetime
 
 import pytz
@@ -65,8 +64,7 @@ class BanjoDriver(object):
         result = results.NdtResult(client=names.BANJO,
                                    start_time=datetime.datetime.now(pytz.utc))
 
-        with contextlib.closing(browser_client_common.create_browser(
-                self._browser)) as driver:
+        with browser_client_common.create_browser(self._browser) as driver:
             result.browser = self._browser
             result.browser_version = driver.capabilities['version']
 
--- a/client_wrapper/browser_client_common.py
+++ b/client_wrapper/browser_client_common.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import contextlib
+
 from selenium import webdriver
 from selenium.common import exceptions
 from selenium.webdriver.support import expected_conditions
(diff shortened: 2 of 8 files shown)


stephen-soltesz:
When I started this review, the build reported failure here: https://github.com/m-lab/ndt-e2e-clientworker/blob/master/tests/test_http_server.py#L61

I reran the travis check three times, and it passed two of three times; I left the third in a failed state. Is that expected?

The issue appears unrelated to these changes. Is this a flaky test?

:lgtm:

---

Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[tests/ndt_client_test.py, line 22 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/55#-KGy1T4_t_vOb0Z-Y18R:-KGy1T4adoa_S4pmB0Qw:-395745088) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/bca742f0720dd95a62ad110adf5de89fea39913a/tests/ndt_client_test.py#L22)):*
nit: and optional, and out of scope for this PR. Based on the file name and this class name, I expected this to be a test, but it's a base class for other ndt client tests. My thought is that the name `NdtClientTestCase` preserves the sense that it's a test case. The file name too, `ndt_client_testcase.py` or `ndt_client_unittest` might help distinguish this file more clearly from an actual test.

Edit: though I see that the file name convention is different than I expect; test files start with `test_` rather than end with `_test`. So, the file name may be clear in context.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/55#-:-KGy07hdwg3ebKCot1QI:-1181740569)*

mtlynch:
> I reran the travis check three times, and it passed two of three times; I left the third in a failed state. Is that expected?

> The issue appears unrelated to these changes. Is this a flaky test?

Yeah, this is #43. It turns out my idea for fixing a race condition by introducing a smaller race condition didn't completely work. I have an idea for a complete elimination of the race, but I've been deferring until we get through the first run of M1 testing.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[tests/ndt_client_test.py, line 22 \[r1\]](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/55#-KGy1T4_t_vOb0Z-Y18R:-KH0LTcCe0peYnSGnEe8:314767562) ([raw file](https://github.com/m-lab/ndt-e2e-clientworker/blob/bca742f0720dd95a62ad110adf5de89fea39913a/tests/ndt_client_test.py#L22)):*
Sure, I like that name better. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/55#-:-KH0Ld9knf9QFb55sjms:696116225)*

PR #64, opened 2016-05-10
Fixing a bug in TestError timestamp calculation

### LINKED ISSUE #60: TestErrors are sharing timestamp
There's an error in the TestError constructor:

https://github.com/m-lab/ndt-e2e-clientworker/blob/master/client_wrapper/results.py#L61

```python
 def __init__(self, message, timestamp=datetime.datetime.now(pytz.utc)):
        self._message = message
        self._timestamp = timestamp
```

This causes the `timestamp` field to be global and all the `TestErrors` share it.

mtlynch:
This fixes a bug in the constructor for TestError. It turns out that the
default value of a function argument is an object, all callers share that
same object. This is undesirable for the TestError because each call should
result in a new timestamp of the current time.

This fixes the shared object and adds a unit test for the issue.

This fixes #60.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/64)

### DIFF
--- a/client_wrapper/results.py
+++ b/client_wrapper/results.py
@@ -58,9 +58,12 @@ class TestError(object):
         timestamp: Datetime of when the error was observed.
     """
 
-    def __init__(self, message, timestamp=datetime.datetime.now(pytz.utc)):
+    def __init__(self, message, timestamp=None):
         self._message = message
-        self._timestamp = timestamp
+        if timestamp:
+            self._timestamp = timestamp
+        else:
+            self._timestamp = datetime.datetime.now(pytz.utc)
 
     def __eq__(self, other):
         return all(((self.message == other.message),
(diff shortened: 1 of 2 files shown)


fernandalavalle:
:lgtm:

<details><summary><i>Previously, coveralls wrote…</i></summary>
> 
> [Coverage Status](https://coveralls.io/builds/6120339)
> 
> Coverage increased (+0.04%) to 88.372% when pulling **dcfd69583268f9ade960e4a0e1d211cf6c0df0cc on mtlynch:fix-error-timestamp** into **3f03ccae949d5f83670a4764626ced2d1926479a on m-lab:master**.
> </details>


---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/64#-:-KHWgfRe536eL1PxGbAD:1417088721)*

PR #77, opened 2016-06-09
Adds a timeout to page load. 

fernandalavalle:
Adds a 10 second timeout to page loads and corresponding tests. 

Timeout does not work for Chrome.


nkinkade:
LGTM

Issue #68, opened 2016-05-27
Banjo - check for start test button throws exception in Chrome+Win10


mtlynch:
I just caught an unexpected exception for the banjo driver:

```
Traceback (most recent call last):
  File "C:\ndt-e2e-clientworker\client_wrapper/client_wrapper.py", line 142, in <module>
    main(parser.parse_args())
  File "C:\ndt-e2e-clientworker\client_wrapper/client_wrapper.py", line 44, in main
    _run_test_iterations(driver, args.iterations, args.output)
  File "C:\ndt-e2e-clientworker\client_wrapper/client_wrapper.py", line 83, in _run_test_iterations
    result = driver.perform_test()
  File "C:\ndt-e2e-clientworker\client_wrapper\banjo_driver.py", line 80, in perform_test
    result).complete_ui_flow()
  File "C:\ndt-e2e-clientworker\client_wrapper\banjo_driver.py", line 103, in complete_ui_flow
    if not self._click_run_test_button():
  File "C:\ndt-e2e-clientworker\client_wrapper\banjo_driver.py", line 151, in _click_run_test_button
    'lrfactory-internetspeed__test_button')
  File "C:\Python27\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 266, in find_element_by_id
    return self.find_element(by=By.ID, value=id_)
  File "C:\Python27\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 744, in find_element
    {'using': by, 'value': value})['value']
  File "C:\Python27\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 233, in execute
    self.error_handler.check_response(response)
  File "C:\Python27\lib\site-packages\selenium\webdriver\remote\errorhandler.py", line 194, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.NoSuchElementException: Message: Unable to locate element: {"method":"id","selector":"lrfactory-internetspeed__test_button"}
Stacktrace:
    at FirefoxDriver.prototype.findElementInternal_ (file:///c:/users/mlab/appdata/local/temp/tmpplfvph/extensions/fxdriver@googlecode.com/components/driver-component.js:10770)
    at FirefoxDriver.prototype.findElement (file:///c:/users/mlab/appdata/local/temp/tmpplfvph/extensions/fxdriver@googlecode.com/components/driver-component.js:10779)
[...]

PR #78, opened 2016-06-16
Banjo RunTest Button 

fernandalavalle:
Added an explicit wait in the banjo driver for waiting until the "Run Test" button is visible before attempting to click it. Allows the driver to gracefully handle not finding the button. 

Addresses this issue: 
https://github.com/m-lab/ndt-e2e-clientworker/issues/68

### DIFF
--- a/client_wrapper/banjo_driver.py
+++ b/client_wrapper/banjo_driver.py
@@ -30,8 +30,6 @@ logger = logging.getLogger(__name__)
 
 ERROR_FAILED_TO_LOCATE_RUN_TEST_BUTTON = (
     'Failed to locate "Run Speed Test" button.')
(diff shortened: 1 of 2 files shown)


nkinkade:
LGTM.  Though, as a separate issue, I wonder whether the _DEFAULT_TIMEOUT of 20 seconds is too high. This E2E testing will happen in a controlled environment (testbed) and if a page element hasn't showed up in the DOM after just a few seconds, I'd say something larger is wrong. I might considering lower that to something more like 5s, unless you and Michael have specific experience showing that waiting longer is better.

fernandalavalle:
I have had no such experience. I had a similar thought when I saw that the default was 20 seconds, but that was out of scope of this PR. I'll ask Michael when he gets back what he thinks about this.

PR #84, opened 2016-06-30
Removing client_url parameter

mtlynch:
This parameter is actually unnecessary. It was originally added for the HTML5
client, which needs a URL rather than a server, but the URL of the HTML5
client can be deduced, given the server FQDN.

Eliminating the client_url parameter simplifies semantics and allows similar
parameters to be used across different client types.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/ndt-e2e-clientworker/84)


fernandalavalle:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/ndt-e2e-clientworker/84#-:-KLXnTj197gFruTBe921:1417088721)*

PR #87, opened 2016-09-27
Adds a successful OneBox reply file generated from google.com

nkinkade:
Adds a successful OneBox reply file generated from google.com, and updates the README with a bit more information about generating future OneBox replay files.


mtlynch:
It might be more convenient to put the replay file in https://github.com/m-lab/ndt-e2e-ansible repo because that's what puts it on the test machines for E2E testing (it looks on the control machine for a file at `{{ local_http_replay_dir }}/{{ http_replay_file }}`.

fernandalavalle:
LGTM

PR #88, opened 2016-09-27
Altered the banjo client path to work with the new replay file. 

fernandalavalle:
Changed the path we run tests against on the local replay server. 

Also in an unrelated change I made output a required argument because when I was testing the new replay file, an error would occur when the output path was not specified.


nkinkade:
LGTM

PR #90, opened 2016-11-23
Updated instructions on generating a replay file to be more detailed

critzo:
This PR adds content to the replay generator readme so it is more specific.

### DIFF
--- a/replay_generator/README.md
+++ b/replay_generator/README.md
@@ -70,30 +70,33 @@ would be saved as:
       ```
 
 ## Example
-To capture traffic for an NDT client at
-http://www.example.com/foo/ndt:
+
+This example assumes you've cloned this repository and installed all requirements using ```pip install -i requirements.txt```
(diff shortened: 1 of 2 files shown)


replay_generator/README.md
```diff
@@ -70,30 +70,33 @@ would be saved as:
...
     ```bash
     python replay_generator/replay_generator.py \
       --port 8888 \
       --output example.com-replay.yaml
     ```
 
-1. Run Firefox in private mode against the target client URL:
+1. Then open a second terminal window and run Firefox in private mode against the target client URL:
     ```bash
     firefox -private http://www.example.com/foo/ndt
```
nkinkade: For what it's worth, my firefox client has no option `-private`, but instead has `--private-window`.

nkinkade:
One small comment about the options passed to Firefox, but otherwise LGTM.

PR #93, opened 2017-02-16
Updates one unit test to check the MacOS version number for Sierra instead of Capitan.

nkinkade:
We've upgraded the MacOS client in the testbed from version Capitan to Sierra. This PR fixes the only reference to "capitan" that I could find in this repository.


fernandalavalle:
LGTM

